PR #26, opened 2015-10-30
update "connection" package to latest 1.0.1 version

mindreframer:
just bumps the version...


michalmuskala:
Since the stable 1.0 is out, I think that relaxing the version requirement to `~> 1.0` would be even better.

mindreframer:
@michalmuskala true. gimme a sec...

PR #104, opened 2017-02-08
feature: find_one

tomjoro:
#29 

Mongodb protocol only has find. find_one is done by just taking the first document using natural order returned from find (as per mongo documentation).

I'm not sure about the applicability of all the options.

### DIFF
--- a/lib/mongo.ex
+++ b/lib/mongo.ex
@@ -319,7 +319,6 @@ defmodule Mongo do
     bangify(distinct(conn, coll, field, filter, opts))
   end
 
-
   @doc """
   Selects document a single document in a collection and returns either a document
   or nil.
@@ -350,7 +349,6 @@ defmodule Mongo do
     |> Enum.to_list
     |> List.first
   end
-
 
   @doc """
   Selects documents in a collection and returns a cursor for the selected
--- a/lib/mongo.ex
+++ b/lib/mongo.ex
@@ -319,6 +319,58 @@ defmodule Mongo do
     bangify(distinct(conn, coll, field, filter, opts))
   end
 
+
+  @doc """
+  Selects documents in a collection and returns a cursor for the selected
+  documents.
+
+  If multiple documents satisfy the query, this method returns the first document
+  according to the natural order which reflects the order of documents on the disk.
+
+  ## Options
+
+    * `:comment` - Associates a comment to a query
+    * `:cursor_type` - Set to :tailable or :tailable_await to return a tailable
+      cursor
+    * `:max_time` - Specifies a time limit in milliseconds
+    * `:modifiers` - Meta-operators modifying the output or behavior of a query,
+      see http://docs.mongodb.org/manual/reference/operator/query-modifier/
+    * `:cursor_timeout` - Set to false if cursor should not close after 10
+      minutes (Default: true)
+    * `:projection` - Limits the fields to return for all matching document
+    * `:skip` - The number of documents to skip before returning (Default: 0)
+  """
+  @spec find_one(conn, collection, BSON.document, Keyword.t) :: cursor
+  def find_one(conn, coll, filter, opts \\ []) do
+    query = [
+      {"$comment", opts[:comment]},
+      {"$maxTimeMS", opts[:max_time]},
+      {"$orderby", opts[:sort]}
+    ] ++ Enum.into(opts[:modifiers] || [], [])
+
+    query = filter_nils(query)
+
+    query =
+      if query == [] do
+        filter
+      else
+        filter = normalize_doc(filter)
+        filter = if List.keymember?(filter, "$query", 0), do: filter, else: [{"$query", filter}]
+        filter ++ query
+      end
+
+    select = opts[:projection]
+    opts = if Keyword.get(opts, :cursor_timeout, true), do: opts, else: [{:no_cursor_timeout, true}|opts]
+    drop = ~w(comment max_time modifiers sort cursor_type projection cursor_timeout order_by)a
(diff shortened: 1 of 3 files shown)


lib/mongo.ex
```diff
@@ -319,6 +319,58 @@ defmodule Mongo do
...
+    opts = if Keyword.get(opts, :cursor_timeout, true), do: opts, else: [{:no_cursor_timeout, true}|opts]
+
+    drop = ~w(comment max_time modifiers sort cursor_type projection cursor_timeout order_by limit)a
+    opts = cursor_type(opts[:cursor_type]) ++ Keyword.drop(opts, drop)
+    opts = [{:limit, 1} | opts]
+
+    cursor = cursor(conn, coll, query, select, opts)
+    list = Enum.to_list(cursor)
+    List.first(list)
+  end
```
ericmj: It looks like we can do

```elixir
opts =
  opts
  |> Keyword.delete(:order_by)
  |> Keyword.put(:limit, 1)
find(conn, coll, filter, opts)
|> Enum.at(0)
```

But please correct me if there are other differences.

lib/mongo.ex
```diff
@@ -319,6 +319,58 @@ defmodule Mongo do
...
+
+
```
ericmj: ✂️ this line.

lib/mongo.ex
```diff
@@ -319,6 +319,58 @@ defmodule Mongo do
     bangify(distinct(conn, coll, field, filter, opts))
   end
 
+
+  @doc """
+  Selects document a single document in a collection and returns either a document
```
ericmj: ✂️ first "document".

lib/mongo.ex
```diff
@@ -319,6 +319,58 @@ defmodule Mongo do
     bangify(distinct(conn, coll, field, filter, opts))
   end
 
+
```
ericmj: ✂️ this line.

test/mongo_test.exs
```diff
@@ -124,6 +124,24 @@ defmodule Mongo.Test do
...
+    assert %{"foo" => 42} = Mongo.find_one(c.pid, coll, %{}, batch_size: 1)
+
+    assert {:ok, _} = Mongo.insert_one(c.pid, coll, %{foo: 43})
+
+    # we should still get one
+    assert %{"foo" => 42} = Mongo.find_one(c.pid, coll, %{}, batch_size: 1)
+
+    assert %{"foo" => 43} != Mongo.find_one(c.pid, coll, %{}, batch_size: 1)
+
+  end
```
ericmj: Since the asserts are still single line we can remove the empty lines around them.

ericmj (changes requested):

lib/mongo.ex
tomjoro: I had also removed the modifier.  Not 100% sure how the modifier works.
      {"$orderby", opts[:sort]}

Enum.at(0) will return "0" if the cursor is so that's why I do the Enum.to_list.

ericmj: You can drop `:sort` from the opts keyword also. `Enum.at(enum, 0)` will return `nil` if the enum is empty.

tomjoro: I guess we could just "find", but would we then replace the entire body of "find_one" with that snippet...

tomjoro:
Enum.at(cursor, 0) doesn't return nil if empty. It returns 0.

ericmj:
> Enum.at(cursor, 0) doesn't return nil if empty. It returns 0.

If it returns `0` then there is an element with value `0` at the first position in the cursor or there is some weird bug in the cursor that would make it return `0` if the cursor returns no results.

```elixir
iex(1)> Enum.at([], 0)
nil
```

tomjoro:
The following is a passing test (which means Enum.at(cursor, 0) returns 0, which is not good).
```
  test "find_is_zero", c do
    coll = unique_name()
    assert 0 == Mongo.find(c.pid, coll, %{}) |> Enum.at(0)
  end
```

ericmj:
The test above fails with the following for me:

```
1) test find_is_zero (Mongo.Test)
     test/mongo_test.exs:142
     Assertion with == failed
     code:  0 == Mongo.find(c.pid(), coll, %{}) |> Enum.at(0)
     left:  0
     right: nil
     stacktrace:
       test/mongo_test.exs:144: (test)
```

What elixir version are you using?

tomjoro:
I'm checking.. I'm on 1.3.2, but looks like I'm on a branch. Sorry. Let me double check.

tomjoro:
Looks like I based on an older branch. My bad. I'll fix this and generate a new proper pull request. I suspect it will work then :)

ericmj:
I bisected and found this commit to fix the bug: https://github.com/elixir-lang/elixir/commit/6f242e7cc27c1956f7f6dece7e38e66a3f2366df. It's a fix we forgot to backport to the v1.3 bugfix branch :(.

> I'll fix this and generate a new proper pull request. I suspect it will work then :)

I don't want to bump the version requirement so keeping your current code is better but please add a note that we should change to Enum.at/2 when we require 1.4.0+.

tomjoro:
You can close/delete this pull request.

I created a new pull requests #105 to get rid of the wrong base for this pull request. But the Enum.at(0) is still returning 0. I wonder if this can be a MongoDB thing?

ericmj:
`Enum.at` was fixed in Elixir 1.4.0, if you use 1.4.0+ you shouldn't get the bug.

You based the PR of the correct branch.

tomjoro:
Ok. I was unsure so I thought best to make a clean pull request, since I locally had some old code which was committed. Can we continue this conversation on #105?

ericmj:
Sure!

PR #127, opened 2017-05-12
Updated the documentation for the Mongo.find/4

BogdanHabic:
As the title suggests the documentation for the `Mongo.find/4` function stated the that `:order_by` option can be passed, but in reality the `:sort` option is used.


ankhers:
Thanks!

PR #130, opened 2017-05-25
add support for responses with 1 or 1.0

zbarnes757:
I was working with Azure's mongodb api and their responses always contain `"ok" => 1` which didn't pattern match so using a soft comparison to make sure both are supported.


ankhers:
Thanks for this @zbarnes757. Would you be able to fix the conflict?

zbarnes757:
@ankhers Should be fixed! Let me know if there is anything else you recommend.

ankhers:
Thanks!

PR #133, opened 2017-06-13
deprecation warning updates for elixir 1.4

matreyes:
- Removing warnings: "please use parentheses to remove the ambiguity or change the variable name"
- Changed mongo default port to 27017 (for testing)

### DIFF
--- a/lib/mongo/topology_description.ex
+++ b/lib/mongo/topology_description.ex
@@ -11,8 +11,8 @@ defmodule Mongo.TopologyDescription do
   alias Mongo.ReadPreference
 
   # see https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#topologydescription
-  @type type :: :"unknown()" | :"single()" | :"replica_set_no_primary()" |
-                :"replica_set_with_primary()" | :"sharded()"
+  @type type :: :unknown | :single | :replica_set_no_primary |
+                :replica_set_with_primary | :sharded
   @type t :: %{
     type: type,
     set_name: String.t | nil,
--- a/lib/mongo/topology_description.ex
+++ b/lib/mongo/topology_description.ex
@@ -11,8 +11,8 @@ defmodule Mongo.TopologyDescription do
   alias Mongo.ReadPreference
 
   # see https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#topologydescription
-  @type type :: :unknown | :single | :replica_set_no_primary |
-                :replica_set_with_primary | :sharded
+  @type type :: :"unknown()" | :"single()" | :"replica_set_no_primary()" |
+                :"replica_set_with_primary()" | :"sharded()"
   @type t :: %{
     type: type,
     set_name: String.t | nil,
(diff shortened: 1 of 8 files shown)


test/mongo/topology_test.exs
```diff
@@ -1,6 +1,6 @@
 defmodule Mongo.TopologyTest do
   use ExUnit.Case # DO NOT MAKE ASYNCHRONOUS
-  alias Mongoman.{ReplicaSet, ReplicaSetConfig}
+  # alias Mongoman.{ReplicaSet, ReplicaSetConfig}
```
ankhers: If there do not need to be aliased, I would prefer it be removed instead of just being commented out.

test/test_helper.exs
```diff
@@ -10,8 +10,8 @@ ExUnit.start()
 
 IO.puts "[mongod v#{version}]"
 
-{_, 0} = System.cmd("mongo", ~w'mongodb_test --eval db.dropDatabase() --port 27001')
-{_, 0} = System.cmd("mongo", ~w'mongodb_test2 --eval db.dropDatabase() --port 27001')
+{_, 0} = System.cmd("mongo", ~w'mongodb_test --eval db.dropDatabase() --port 27017')
+{_, 0} = System.cmd("mongo", ~w'mongodb_test2 --eval db.dropDatabase() --port 27017')
```
ankhers: The ports being 27001 was actually here because of the three mongo instances that get started automatically when running the test suite (See `start_mongo.bash`).

I will need to see if these lines are still needed. I also wanted to move the starting of mongo inside this file so that it will automatically close upon test completion instead of keeping the three extra servers around.

ankhers:
ping?

matreyes:
I'm sorry... pong!

lib/mongo/topology_description.ex
```diff
@@ -11,8 +11,8 @@ defmodule Mongo.TopologyDescription do
   alias Mongo.ReadPreference
 
   # see https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#topologydescription
-  @type type :: :unknown | :single | :replica_set_no_primary |
-                :replica_set_with_primary | :sharded
+  @type type :: :"unknown()" | :"single()" | :"replica_set_no_primary()" |
+                :"replica_set_with_primary()" | :"sharded()"
```
ankhers: Why are you adding parenthesis to these? They are just atoms, not functions.

matreyes: You are right.
I don't know why the compiler was checking this typespec on the first argument of `TopologyDescription.select_servers/3`

And, it was calling a function on 
[test/mongo/topology_description_test.rb](https://github.com/ankhers/mongodb/blob/master/test/mongo/topology_description_test.exs) 
`TopologyDescription.select_servers(single(), :read, opts)`

Shouldn't be set as private `@typep type`?

I'm sorry, I'm too newbie on this, and thanks for the patience!

ankhers: In that instance, the `single/0` function comes from `Mongo.TopologyTestData`. Also, the Elixir compiler does not check types.

matreyes: Yes!.. but I got some errors... and now I don't... really don't know what was happening, but now it seems to work well.
Thanks!

ankhers:
Closing in favour of #191

Issue #142, opened 2017-07-17
A single mongo node being brought down will crash the entire app


ben-ic:
If you are using mongodb replica set and one node becomes unavailable the entire mongodb driver will crash ultimately crashing the entire app.

ankhers:
So this was actually a misfeature of the `db_connection` package. This has been resolved in master and will released with v1.1. Unfortunately I do not know when it will actually be released. 

I have offered to help expedite the process, so I am hoping to have my portion of that work done over the weekend.

christianjgreen:
Isn't `db_connection` already on 1.1.2 as of march?

ankhers:
You are right. I missed that originally. However, the rest of what I said is still true. The issue has been fixed on master and will be released hopefully somewhat shortly.

christianjgreen:
Just brought in master branch of `db_connection` and still had the application shutdown after one of the 3 secondary nodes in the replica set went down.

ankhers:
Can you please post the error you received?

christianjgreen:
The majority of the errors were
```
 (CaseClauseError) no case clause matching: {:error, %DBConnection.ConnectionError{message: "connection not available because of disconnection"}}
    (mongodb) lib/mongo/monitor.ex:94: Mongo.Monitor.is_master/3
    (mongodb) lib/mongo/monitor.ex:70: Mongo.Monitor.check/1
    (stdlib) gen_server.erl:601: :gen_server.try_dispatch/4
    (stdlib) gen_server.erl:667: :gen_server.handle_msg/5
    (stdlib) proc_lib.erl:247: :proc_lib.init_p_do_apply/3
Last message: {:EXIT, #PID<0.588.0>, {{:case_clause, {:error, %DBConnection.ConnectionError{message: "connection not available because of disconnection"}}}, [{Mongo.Monitor, :is_master, 3, [file: 'lib/mongo/monitor.ex', line: 94]}, {Mongo.Monitor, :check, 1, [file: 'lib/mongo/monitor.ex', line: 70]}, {:gen_server, :try_dispatch, 4, [file: 'gen_server.erl', line: 601]}, {:gen_server, :handle_msg, 5, [file: 'gen_server.erl', line: 667]}, {:proc_lib, :init_p_do_apply, 3, [file: 'proc_lib.erl', line: 247]}]}}
[...]

Issue #141, opened 2017-07-16
How to deal with database going down?


elie222:
If we the database is shut off, Elixir will crash after retrying 5 times in 5 seconds. How we have things at the moment is that it will bring down our entire application.

This may be alright for us. I'm not sure how much we can do when our database is down. Question is how do we stick things back up when this happens? What's the recommended way to check for restart a minute later to see if mongo connection is back?

michalmuskala:
This sounds like there's some fault in the driver. `db_connection` itself is generally geared towards an incremental backoff to reconnect instead of crashing in case of failure. What processes exactly are crashing? I'm suspecting there might be something up with the new topology thingy since it's outside `db_connection` itself.

ben-ic:
I was able to pinpoint where the problem is in the mongodb package.

In lib/mongo/protocol.ex not all possible disconnect reasons are taken into account
```ruby
defp connect(opts, s) do
    ...
    case result do
      {:ok, s} ->
        {:ok, s}
      {:disconnect, {:tcp_recv, reason}, _s} ->
        {:error, Mongo.Error.exception(tag: :tcp, action: "recv", reason: reason)}
      {:disconnect, {:tcp_send, reason}, _s} ->
        {:error, Mongo.Error.exception(tag: :tcp, action: "send", reason: reason)}
      {:disconnect, %{code: code, message: message}, _s} ->                                   #NEEDS TO BE ADDED                 
        {:error, Mongo.Error.exception(tag: :tcp, action: "recv", reason: message)}
      {:error, reason} ->
        {:error, reason}
    end
  end
```


In lib/mongo.ex we call the DBConnection.execute function numerous times.
When mongodb is not connected DBConnection.execute raises an error, which is not being caught, which causes the supervisor to crash.  This may be by design, but this is problematic as it is causing the entire app to crash when mongodb is disconnected.
This leaves us with two other possibilities that I can think of,
1. Adding a try/rescue block around DBConnection.execute (not very elixir like), but something similar to the code below would work.
```ruby
def direct_command(conn, query, opts \\ []) do
    params = [query]
    query = %Query{action: :command}
    try do
      with {:ok, reply} <- DBConnection.execute(conn, query, params,
                                                defaults(opts)) do
        case reply do
          op_reply(flags: flags, docs: [%{"$err" => reason, "code" => code}])
              when (@reply_query_failure &&& flags) != 0  ->
            {:error, Mongo.Error.exception(message: reason, code: code)}
[...]

PR #145, opened 2017-07-24
WIP: Better replica sets

### LINKED ISSUE #144: Application will not start if one or more servers in the cluster are down
I am not 100% sure what part of the pipeline is broken here. But I think we should make connections to the monitors first, then make the actual connections the application will use based on which servers were connected to successfully.

ankhers:
This PR aims to better follow the [MongoDB server discovery and monitoring spec](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#topologytype)

This is PR aims to resolve #144, #142 and #141. I believe there is overlap between those different issues.

### DIFF
--- a/lib/mongo/events.ex
+++ b/lib/mongo/events.ex
@@ -3,54 +3,54 @@ defmodule Mongo.Events do
   def notify(event) do
     GenEvent.notify(__MODULE__, event)
   end
-end
 
-defmodule Mongo.Events.ServerDescriptionChangedEvent do
-  @moduledoc "Published when server description changes, but does NOT include changes to the RTT."
-  defstruct [:address, :topology_pid, :previous_description, :new_description]
-end
+  defmodule ServerDescriptionChangedEvent do
+    @moduledoc "Published when server description changes, but does NOT include changes to the RTT."
+    defstruct [:address, :topology_pid, :previous_description, :new_description]
+  end
 
-defmodule Mongo.Events.ServerOpeningEvent do
-  @moduledoc "Published when server is initialized."
-  defstruct [:address, :topology_pid]
-end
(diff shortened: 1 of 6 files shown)


ankhers:
@ben-ic, @Arthien, @elie222 Would the three of you mind testing this branch to see if you are still having issues with your applications crashing / not starting when a Mongo node goes down?

christianjgreen:
@ankhers sure!

christianjgreen:
@ankhers Working here so far! Was able to stay connected

ankhers:
Assuming no one mentions something about this not working correctly, I will merge it once the new version of the `db_connection` package is released.

christianjgreen:
@ankhers what changes do you need from `db_connection`?

ankhers:
@Arthien It has to do with how a not connected connection worked. In the current released version if you attempt to query against a not connected connection, it would raise an exception. This has been changed to return `{:error, exception}` instead. This was one of the issues that you were bumping up against when one or more servers in the MongoDB cluster were down.

This is important because we keep trying to query each of the servers in the cluster to see what its representation of the cluster looks like so that the driver can keep up to date on how things should be looking.

ankhers:
@Arthien Can you test to see if this works for you still?

ankhers:
ping @Arthien

christianjgreen:
@ankhers I'll test this out today!

ankhers:
Thanks!

christianjgreen:
@ankhers Looking good. Scaling a 3 cluster replica set up and then down. Application did not restart and I was able to stay connected to at least one node.

ankhers:
Good to hear. I will create a release for this branch shortly.

PR #173, opened 2017-12-21
Add support for mongoUrl

sauvainr:
While starting to use this Driver in our company, the missing mongo url support is a pain so here is the parser I used on our application side.
Hopefully we can have it integrated in the driver.


ankhers:
Thank you for this!

sauvainr:
Thanks

PR #174, opened 2017-12-21
Add admin_command function

sauvainr:
Adding a helper function to do admin command operation as creating or dropping collections.

### DIFF
--- a/lib/mongo.ex
+++ b/lib/mongo.ex
@@ -445,6 +445,16 @@ defmodule Mongo do
          do: direct_command(conn, query, opts)
   end
 
+  @doc """
+  Issue a database admin command.
+  List of admin commands: https://docs.mongodb.com/manual/reference/command/
+  """
+  @spec admin_command(GenServer.server, BSON.document, Keyword.t) :: result(BSON.document)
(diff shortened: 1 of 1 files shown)


ankhers:
What does this add that the current `Mongo.command/3` does not already do?

sauvainr:
Mostly it targets `:write` instead of `:read` which I suppose force to target the primary node.
I havn't tested with :read while preference are set to `:secondary` however it seems a bit odd.
Maybe the original command could target `:write` ?

ankhers:
Right. I missed that the first time. I am not sure if we can switch it to target write. Maybe we can default to `:read` and give an option to switch to `:write` if necessary.

Let me think about this for a little while.

ankhers:
After going through the code again, I don't think this is necessary. The current `Mongo.command/3` function defaults to read preference `:primary`, so it should hit the primary node unless you specifically tell it that it is okay to hit a secondary.

I just tested this and it worked as described above. I created and deleted a user giving no `:read_preference` option and it worked. I then tried to create a user passing in a `:read_preference` stating that I want it to hit `:secondary` and it failed because it cannot write to a non-master.

PR #180, opened 2018-01-03
Map maxPoolSize Url parameter

sauvainr:
Mapping mongodb Url to DBConnection.poolboy option.

See: https://github.com/elixir-ecto/db_connection/blob/master/lib/db_connection/poolboy.ex#L79


ankhers:
Would you please be able to add a test for this?

ankhers:
Thank you!

PR #190, opened 2018-01-25
Fix Mongo.find_one/3 typespec

aaronrenner:
`Mongo.find_one/3` returns a BSON document or nil if nothing is found, not a `Mongo.Cursor`.


ankhers:
Thank you!

PR #191, opened 2018-01-26
Elixir 1.6

pragdave:
Here are the changes that remove warnings from Elixir 1.6 compilations.

The vast majority were trivial.

The change to the Cursors (adding `slice/1` which is now required by Enumerable) should be 100% backwards compatible.

The biggest issue was the deprecation of GenEvent. However, as far as I can tell this was only used in one place, waiting for a topology change. So I added a trivial genserver to Mongo.Events which basically does event dispatch, along with a function that waits for a given time for a particular event to be notified. 

I wasn't too sure if the GenEvent was also an external service API, in which case my change is wrong. However, something will need to be done in this case, because :gen_event doesn't know about streams.

Also, I ran Dialyzer and got 16 warnings. Some I suspect are bogus, but many (such as cr:23) appear genuine. Just an FYI:

~~~
:0: Unknown function ssl:connect/3
:0: Unknown function ssl:format_error/1
:0: Unknown function ssl:getopts/2
:0: Unknown function ssl:setopts/2
:0: Unknown type 'Elixir.BSON.DateTime':t/0
:0: Unknown type 'Elixir.Mongo.Error':t/0
lib/mongo.ex:175: Function find_one_and_update/4 has no local return
lib/mongo.ex:175: Function find_one_and_update/5 has no local return
lib/mongo.ex:190: The call 'Elixir.Keyword':drop(Vopts@1::'nil' | [{atom(),_}] | map(),[<<_:32,_:_*8>>,...]) breaks the contract (t(),[key()]) -> t()
lib/mongo.ex:215: Function find_one_and_replace/4 has no local return
lib/mongo.ex:215: Function find_one_and_replace/5 has no local return
lib/mongo.ex:230: The call 'Elixir.Keyword':drop(Vopts@1::'nil' | [{atom(),_}] | map(),[<<_:32,_:_*8>>,...]) breaks the contract (t(),[key()]) -> t()
lib/mongo.ex:251: Function find_one_and_delete/3 has no local return
lib/mongo.ex:251: Function find_one_and_delete/4 has no local return
lib/mongo.ex:261: The call 'Elixir.Keyword':drop(Vopts@1::'nil' | [{atom(),_}] | map(),[<<_:32,_:_*8>>,...]) breaks the contract (t(),[key()]) -> t()
lib/mongo.ex:310: Function distinct/4 has no local return
lib/mongo.ex:310: Function distinct/5 has no local return
lib/mongo.ex:318: The call 'Elixir.Keyword':drop(Vopts@1::'nil' | [{atom(),_}] | map(),[<<_:64>>,...]) breaks the contract (t(),[key()]) -> t()
lib/mongo.ex:329: Function 'distinct!'/4 has no local return
lib/mongo.ex:329: Function 'distinct!'/5 has no local return
lib/mongo.ex:916: Guard test is_map(Vdoc@1::nonempty_maybe_improper_list()) can never succeed
lib/mongo.ex:922: The call 'Elixir.Mongo':assign_id(Vdoc@1::nonempty_maybe_improper_list()) will never return since it differs in the 1st argument from the success typing arguments: (map())
lib/mongo/auth/cr.ex:23: The pattern Verror@1 = {'error', _} can never match the type {'ok',_} | {'disconnect',#{'__exception__':='true', '__struct__':='Elixir.Mongo.Error', 'code':=_, 'message':=_},#{'socket':={_,_}, _=>_}}
lib/mongo/auth/scram.ex:29: The pattern Verror@1 = {'error', _} can never match the type {'ok',_} | {'disconnect',#{'__exception__':='true', '__struct__':='Elixir.Mongo.Error', 'code':=_, 'message':=_},#{'socket':={_,_}, _=>_}}
lib/mongo/messages.ex:135: The variable _@1 can never match since previous clauses completely covered thetype <<_:64>>
lib/mongo/messages.ex:135: Cons will produce an improper list since its 2nd argument is <<_:64>>
lib/mongo/protocol.ex:165: The call inet:setopts(Vsock@1::any(),[{'buffer',<<>>},...]) breaks the contract (Socket,Options) -> 'ok' | {'error',posix()} when Socket :: socket(), Options :: [socket_setopt()]
lib/mongo/protocol/utils.ex:130: The call 'Elixir.Mongo.Protocol.Utils':assign_id(Vdoc@1::nonempty_maybe_improper_list()) will never return since it differs in the 1st argument from the success typing arguments:(map())~~~


ankhers:
Thank you for your work on this. Unfortunately I am away for a few more
days, so I will review these changes when I get back to a solid internet
service.

ankhers:
Thank you again for working on this. However, I would be more comfortable with the `GenEvent` changes being made in a different PR. Is it possible to revert those changes and we can figure out the best strategy for that change?

pragdave:
I _think_ I have it reverted.

Dave

ankhers:
Thank you for these changes. Unfortunately I believe you jumped the gun on the child spec changes. This library still supports Elixir 1.3+ at the moment. Once that is fixed, I would be happy to merge this PR!

ankhers:
Thank your for the contributions!

PR #196, opened 2018-02-19
Fix for trying to query the connection too soon

### LINKED ISSUE #158: after_connect not populating Topology.state.connection_pools in time
I'm using poolboy to start workers that make connections to Mongo.
If I send too many requests to poolboy and it spawns up new connections to Mongo, there is a timing issue where `state.connection_pools` in `Topology.ex` isn't populated in time because it receives messages out of order.

For example:

Mongo.ex
```
@spec start_link(Keyword.t) :: {:ok, pid} | {:error, Mongo.Error.t | term}
def start_link(opts) do
  res = Topology.start_link(opts)
  IO.inspect(res, label: "Started Mongo")
  res
end

@spec find(GenServer.server, collection, BSON.document, Keyword.t) :: cursor
def find(topology_pid, coll, filter, opts \\ []) do
  IO.inspect(topology_pid, label: "Finding")
  ...
end
```

Monitor.ex
```
@doc false
def connected(_connection, me, topology_pid) do
  IO.inspect(topology_pid, label: "Connected")
  GenServer.cast(topology_pid, {:connected, me})
end
```

Topology.ex
```
def handle_call({:connection, address}, _from, state) do
  IO.inspect(state.connection_pools, label: "Connection Pools #{inspect(self)}")
  {:reply, Map.fetch(state.connection_pools, address), state}
end
```

I will get output that looks like this:

```
Started Mongo: {:ok, #PID<0.2690.0>}
Finding: #PID<0.2690.0>
Connection Pools #PID<0.2690.0>: %{}
Connected: #PID<0.2690.0>
```

I believe the `after_connect` in DbConnection is handled after a pid is returned and poolboy starts using the pid before the `connected` is called in Topology which would result in an inbox for Topology looking like this:

```
:connection // called from connection_for_address
:connected
```

I recommend Monitor.init/1 be the following:

```
@doc false
  def init([server_description, topology_pid, heartbeat_frequency_ms, connection_opts]) do
    opts = # monitors don't authenticate and use the "admin" database
      connection_opts
      |> Keyword.put(:database, "admin")
      |> Keyword.put(:skip_auth, true)
      |> Keyword.put(:backoff_min, heartbeat_frequency_ms)
      |> Keyword.put(:backoff_max, heartbeat_frequency_ms)
      |> Keyword.put(:backoff_type, :rand)
      |> Keyword.put(:connection_type, :monitor)
      |> Keyword.put(:topology_pid, topology_pid)

    {:ok, pid} = DBConnection.start_link(Mongo.Protocol, opts)
    
    # Notify Topology that we are connected
    connected(pid, self(), topology_pid)
    
    :ok = GenServer.cast(self, :check)
    {:ok, %{
      connection_pid: pid,
      topology_pid: topology_pid,
      server_description: server_description,
      heartbeat_frequency_ms: heartbeat_frequency_ms,
      opts: opts
    }}
  end
```

Here we drop `after_connect` and call `connected` right away.

If you agree that I've identified the problem and accept the proposed solution, I can submit a pull request.

ankhers:
Fixes #158

### DIFF
--- a/lib/mongo.ex
+++ b/lib/mongo.ex
@@ -739,15 +739,20 @@ defmodule Mongo do
 
   defp select_servers(topology_pid, type, opts) do
     start_time = System.monotonic_time
-    _select_servers(topology_pid, type, opts, start_time)
+    select_servers(topology_pid, type, opts, start_time)
   end
 
   @sel_timeout 30000
-  defp _select_servers(topology_pid, type, opts, start_time) do
+  defp select_servers(topology_pid, type, opts, start_time) do
     topology = Topology.topology(topology_pid)
     with {:ok, servers, slave_ok, mongos?} <- TopologyDescription.select_servers(topology, type, opts) do
       if Enum.empty? servers do
-        Topology.wait_for_connection(topology_pid, @sel_timeout, start_time)
+        case Topology.wait_for_connection(topology_pid, @sel_timeout, start_time) do
+          {:ok, servers} ->
+            select_servers(topology_pid, type, opts, start_time)
+          {:error, :selection_timeout} = error ->
+            error
+        end
       else
         {:ok, servers, slave_ok, mongos?}
       end
--- a/lib/mongo.ex
+++ b/lib/mongo.ex
@@ -48,7 +48,6 @@ defmodule Mongo do
   use Bitwise
   use Mongo.Messages
   alias Mongo.Query
-  alias Mongo.Events.TopologyDescriptionChangedEvent
   alias Mongo.ReadPreference
   alias Mongo.TopologyDescription
   alias Mongo.Topology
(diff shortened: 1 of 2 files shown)


ankhers:
@Sailias, @gazoon Can you please test this branch to ensure that it fixed your issues?

ankhers:
Actually, don't test yet. I still have things failing on my end.

ankhers:
Alright, should be good to test.

ankhers:
@Sailias The repo you linked in your ticket works for me under these changes.

Sailias:
@ankhers does that mean the repo I linked didn’t work prior to the changes as you commented before (but may have deleted)?

ankhers:
As in I was getting the error you mentioned before these changes. With these changes I am no longer getting an error.

PR #199, opened 2018-03-14
fix #197

### LINKED ISSUE #197: Mongo.Protocol still using send_timeout
https://elixirforum.com/t/poolboy-for-mongodb-driver-pool-timeout-seems-not-working/12842/2

We just need to remove the send_timeout option.

shavit:
Remove the `:send_timeout` option from the socket


ankhers:
Thank you very much for this!

PR #200, opened 2018-03-16
Add missing case clause in Mongo.Protocol.connect/2

jonbusby:
`Mongo.Protocol.wire_version/1` can return a `{:disconnect, %Mongo.Error{}, s}`, but `Mongo.Protocol.connect/2` is missing a case clause to handle that. This manifested for me as a crash after shutting down a replica set member while testing replica set failover handling.


ankhers:
Thank you!

PR #206, opened 2018-05-22
remove connection to arbiter when role changes

wild5r:
Issue:
* when server's role changes to `arbiter` respective connection remains in the connection pool
* hence the server no longer authorizes incoming connections

Solution:
* remove server connection from pool once the role gets clear


ankhers:
Thank you!

PR #219, opened 2018-07-18
Support for mongodb+srv url

rafaelm7o:
This is the implementation for supporting mongodb+srv URL
I tested it with a MongoDB Atlas Account-
For the test cases, I simulated the SRV and TXT record with my own domain

### DIFF
--- a/lib/mongo/url_parser.ex
+++ b/lib/mongo/url_parser.ex
@@ -111,10 +111,8 @@ defmodule Mongo.UrlParser do
   defp resolve_srv_url(%{"seeds" => url, "srv" => srv} = frags)
        when is_bitstring(url) and srv == "+srv" do
     # Fix for windows only
-    case :os.type() do
-      {:win32, _} ->
-        :inet_db.add_ns({4, 2, 2, 1})
-      _ ->
+    with {:win32, _} <- :os.type() do
+         :inet_db.add_ns({4, 2, 2, 1})
     end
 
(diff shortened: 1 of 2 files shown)


lib/mongo/url_parser.ex
```diff
@@ -114,6 +114,7 @@ defmodule Mongo.UrlParser do
     case :os.type() do
       {:win32, _} ->
         :inet_db.add_ns({4, 2, 2, 1})
+      _ ->
```
ankhers: I would rather return an explicit value here instead of the implicit `nil`.

ankhers:
I would also rather not rely on your domain for the tests. Maybe we should look into what official drivers are doing for their tests?

rafaelm7o:
The problem was I did not find any test official domain for that, That is the reason for I decided to use the custom domain

ankhers:
* [ruby driver](https://github.com/mongodb/mongo-ruby-driver/search?utf8=%E2%9C%93&q=srv&type=)
* [python driver](https://github.com/mongodb/mongo-python-driver/search?q=srv&unscoped_q=srv)

It looks like there is a number of 10gen.cc domains that they are using.

rafaelm7o:
I changed the URL. I haven't seen the test cases, and certainly, some will fail. 
Ex. with more than one TXT record.

It works in a happy path scenery

ankhers:
Thank you!

PR #221, opened 2018-07-26
✨ query decimal128

chakming:
- support query `decimal128`
- decode the bits into `Decimal`
  - added dependency on `decimal` library


ankhers:
It looks like your changes made the ssl tests run.

chakming:
yes, will keep working on it :)

ankhers:
Sorry for the delay here. Thank you!

PR #222, opened 2018-07-27
Fix unused variable warning in Mongo.Topology

LysanderGG:
Fix the following warning:

```
warning: variable "from" is unused
  lib/mongo/topology.ex:124
```

Sorry if I missed a fix in another branch.


ankhers:
Thank you!

PR #225, opened 2018-08-03
Fix dialyzer

gpad:
Fix some specs and some small bug that dialyzer discover. I added also the dialyzer task on travis only on elixir version 1.6.


ankhers:
There was quite a bit of code that was commented out. Would you be able to remove it instead if it was not used?

gpad:
@ankhers Removed all the dead code ...

gpad:
@ankhers I have update my PR with master let me know WDYT or If I can help you in other way ...

ankhers:
Thanks! I'm just waiting for travis to do it's thing and when it passes I will be merging.

ankhers:
Thank you!

PR #228, opened 2018-08-08
Fix MatchError on missing monitor_pid

tschmittni:
This fixes the following match error:

```
GenServer :mongo terminating
** (MatchError) no match of right hand side value: nil
    (mongodb) lib/mongo/topology.ex:146: Mongo.Topology.handle_cast/2
    (stdlib) gen_server.erl:616: :gen_server.try_dispatch/4
    (stdlib) gen_server.erl:686: :gen_server.handle_msg/6
    (stdlib) proc_lib.erl:247: :proc_lib.init_p_do_apply/3
```

### DIFF
--- a/lib/mongo/topology.ex
+++ b/lib/mongo/topology.ex
@@ -143,26 +143,29 @@ defmodule Mongo.Topology do
   end
 
   def handle_cast({:connected, monitor_pid}, state) do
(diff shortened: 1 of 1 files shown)


ankhers:
Thank you!

Issue #211, opened 2018-06-26
List Collections


caldempsey:
No functionality yet implemented to list collections. As advised this is something I will be looking into this weekend.

ankhers:
The [specification](https://github.com/mongodb/specifications/blob/f4bb783627e7ed5c4095c5554d35287956ef8970/source/enumerate-collections.rst) may help a lot here. Feel free to ping me if you have any questions.

You will also want to check the wire version, if it is less than 3, then you will do [this](https://github.com/mongodb/specifications/blob/f4bb783627e7ed5c4095c5554d35287956ef8970/source/enumerate-collections.rst#pre-mongodb-276-versions) section. If it is greater than or equal to 3, you will do [this](https://github.com/mongodb/specifications/blob/f4bb783627e7ed5c4095c5554d35287956ef8970/source/enumerate-collections.rst#post-mongodb-280-rc3-versions) section. You should be able to find an example of how to check the wireversion in the code.

ankhers:
[This](https://github.com/mongodb/mongo/blob/master/src/mongo/db/wire_version.h) list of wire versions may also come in handy.

ankhers:
This was merged in #240

PR #240, opened 2018-10-18
GridFS support

### LINKED ISSUE #238: Using mongod v3.6.4, test aggregate (Mongo.Test) fails
I have started a mongodb with defaults params and the replica set with the `start_mongo.bash` script. Running
```
mix test --exclude ssl
```

I got the error:
```

  1) test aggregate (Mongo.Test)
     test/mongo_test.exs:48
     ** (CaseClauseError) no case clause matching: {:ok, %{cursor_id: 0, docs: [%{"$clusterTime" => %{"clusterTime" => #BSON.Timestamp<1539546016:6>, "signature" => %{"hash" => #BSON.Binary<0000000000000000000000000000000000000000>, "keyId" => 0}}, "code" => 9, "codeName" => "FailedToParse", "errmsg" => "The 'cursor' option is required, except for aggregate with the explain argument", "ok" => 0.0, "operationTime" => #BSON.Timestamp<1539546016:6>}], from: 0, num: 1}}
     code: assert []               = Mongo.aggregate(c.pid, coll, [], use_cursor: false) |> Enum.drop(4)
     stacktrace:
       (mongodb) lib/mongo/cursor.ex:232: anonymous fn/5 in Enumerable.Mongo.SinglyCursor.start_fun/5
       (elixir) lib/stream.ex:1362: anonymous fn/5 in Stream.resource/3
       (elixir) lib/enum.ex:2979: Enum.drop/2
       test/mongo_test.exs:82: (test)
```

I can fix the test by using the `use_cursor: true`

```
    assert []               = Mongo.aggregate(c.pid, coll, [], use_cursor: true) |> Enum.drop(4)
    assert [%{"foo" => 42}] = Mongo.aggregate(c.pid, coll, [], use_cursor: true) |> Enum.take(1)
    assert [%{"foo" => 45}] = Mongo.aggregate(c.pid, coll, [], use_cursor: true) |> Enum.drop(3)
```

I do not know why now, but I think it's the MongoDB version. Can somebody confirm this?

zookzook:
resolves #238, #211 and #73.

* grid-fs support for uploading and downloading files (streams)
* list collections `Mongo.show_collections/2`
* list indexed for collections `Mongo.list_indexes/3`
* unit tests for grid-fs included

maybe some changes go too far
* update the dialyxir version to 1.0.0-rc.3
* update the elixir version to 1.7
* fixed some deprecated warnings (if not x in y vs if x not in y)


ankhers:
Thank you for your work on this! I have not yet looked over the code, but based on your initial comment, the Elixir 1.7 is too far. The highest I'm willing to go on it right now is 1.4.

Is this considered complete at the moment, or are you still making some changes? Either way, I will start taking a look at things over the next couple days.

zookzook:
Yes. I try to get travis happy. So I changed some lines to run with 1.4 again. But if travis passed then it should be complete.

lib/mongo.ex
```diff
@@ -36,7 +36,7 @@ defmodule Mongo do
   All write operations take the following options for controlling the
   write concern.
 
-    * `:w` - The number of servers to replicate to before returning from write
```
ankhers: I think this was an accidental removal?

lib/mongo.ex
```diff
@@ -790,6 +795,92 @@ defmodule Mongo do
     bangify(update_many(topology_pid, coll, filter, update, opts))
   end
 
+  @doc """
+  Returns a cursor to enumerate all indexes
+
+  :full returns the full index information instead of the name only
```
ankhers: You should add that it defaults to `false`.

lib/mongo.ex
```diff
@@ -790,6 +795,92 @@ defmodule Mongo do
...
+  end
+
+  @doc """
+  Getting Collection Names
+  """
+  @spec show_collections(GenServer.server, Keyword.t) :: cursor
+  def show_collections(topology_pid, opts \\ []) do
+
+    #
+    # todo:
```
ankhers: Are these todo items complete?

zookzook:
It is done!

lib/mongo.ex
```diff
@@ -790,6 +795,92 @@ defmodule Mongo do
...
+    #
+    #
+    wv_query = %Query{action: :wire_version}
+
+    with {:ok, conn, _, _} <- select_server(topology_pid, :read, opts),
+         {:ok, version}    <- DBConnection.execute(conn, wv_query, [], defaults(opts)) do
+
+      case version do
+        v when v > 2 -> show_collections_v3(conn, opts)
+        _            -> show_collections_v2(topology_pid)
```
ankhers: I don't think we need to worry about wire version <= 2. Version 3 was introduced with mongo version 2.7.7, which we don't support. We only support 3.0+ right now.

lib/mongo/grid_fs/bucket.ex
```diff
@@ -0,0 +1,229 @@
...
+             }
+
+  defstruct name: "fs", chunk_size: 255 * 1024, topology_pid: nil
+
+  @doc """
+  Creates a new Bucket with a existing connection using the default values. It just contains the
+  name of the collections (fs) and the chunk size (255KB).
+
+  The bucket checks the index for both collections as well. In case of multiple
+  upload or downloads just create only one bucket and resuse it.
```
ankhers: Typo: "resuse".

lib/mongo/grid_fs/bucket.ex
```diff
@@ -0,0 +1,229 @@
...
+  @spec rename(Bucket.t, BSON.ObjectId.t, String.t) :: Mongo.result(BSON.document)
+  def rename(%Bucket{topology_pid: topology_pid} = bucket, file_id, new_filename) do
+    query       = %{_id: file_id}
+    update      = %{"$set" => %{filename: new_filename}}
+    collection  = files_collection_name(bucket)
+    {:ok, _doc} = Mongo.find_one_and_update(topology_pid, collection, query, update)
+  end
+
+  @doc """
+  Given a @id, delete this stored file’s files collection document and
```
ankhers: I think this should read "Given an \`id\`

lib/mongo/grid_fs/download.ex
```diff
@@ -0,0 +1,80 @@
...
+  specified by fileId.
+  Returns a Stream.
+  """
+  @spec open_download_stream(Bucket.t, String.t | BSON.ObjectId.t | map()) :: result
+  def open_download_stream(bucket, file_id) when is_binary(file_id) do
+    find_one_file(bucket, %{"_id" => ObjectId.decode!(file_id)})
+  end
+
+  @doc """
+  Same as above, accepting an OID
```
ankhers: Don't say "Same as above". This is all the same function. Document it as a single function.

mix.lock
```diff
@@ -1,10 +1,22 @@
...
+  "parse_trans": {:hex, :parse_trans, "3.3.0", "09765507a3c7590a784615cfd421d101aec25098d50b89d7aa1d66646bc571c1", [:rebar3], [], "hexpm"},
   "poolboy": {:hex, :poolboy, "1.5.1", "6b46163901cfd0a1b43d692657ed9d7e599853b3b21b95ae5ae0a777cf9b6ca8", [:rebar], [], "hexpm"},
+  "ssl_verify_fun": {:hex, :ssl_verify_fun, "1.1.4", "f0eafff810d2041e93f915ef59899c923f4568f4585904d010387ed74988e77b", [:make, :mix, :rebar3], [], "hexpm"},
+  "timex": {:hex, :timex, "3.4.1", "e63fc1a37453035e534c3febfe9b6b9e18583ec7b37fd9c390efdef97397d70b", [:mix], [{:combine, "~> 0.10", [hex: :combine, repo: "hexpm", optional: false]}, {:gettext, "~> 0.10", [hex: :gettext, repo: "hexpm", optional: false]}, {:tzdata, "~> 0.1.8 or ~> 0.5", [hex: :tzdata, repo: "hexpm", optional: false]}], "hexpm"},
+  "tzdata": {:hex, :tzdata, "0.5.19", "7962a3997bf06303b7d1772988ede22260f3dae1bf897408ebdac2b4435f4e6a", [:mix], [{:hackney, "~> 1.0", [hex: :hackney, repo: "hexpm", optional: false]}], "hexpm"},
+  "unicode_util_compat": {:hex, :unicode_util_compat, "0.4.1", "d869e4c68901dd9531385bb0c8c40444ebf624e60b6962d95952775cac5e90cd", [:rebar3], [], "hexpm"},
```
ankhers: There seems to be a lot of unnecessary things added to the lock file. Can you please clean this up?

test/mongo_test.exs
```diff
@@ -28,6 +28,23 @@ defmodule Mongo.Test do
     end
   end
 
+#  test "show_collections", c do
```
ankhers: What was the purpose of commenting out this test?

test/test_helper.exs
```diff
@@ -17,6 +17,7 @@ options = if version < {3, 4, 0} do [mongo_3_4: true] ++ options else options en
 ExUnit.configure exclude: options
 ExUnit.start()
 
+### currently disabled
```
ankhers: What exactly is disabled?

ankhers (commented):
I quickly went through the code and had a couple questions and suggested changes. I will go through it more throughly when I have a bit more time.

lib/mongo.ex
zookzook: Yes, it was.

lib/mongo.ex
zookzook: Ok, I will remove support for < 3.0

test/test_helper.exs
zookzook: I removed the comment. I started the test with a local mongodb instance and comment some lines in the test_helper.exs.

test/mongo_test.exs
zookzook: Ok. I forgot this test. I will fix it.

zookzook:
zookzook commented 25 days ago....is this means: a couple of days

    A "couple of days' can be interpreted as is or can mean three or four days. Several people can
    interpret the phrase literally or loosely. It all depends on what the person saying it means. ... You
    can simple ask the person exactly what is meant when he/she tells you "couple of days"for a clear
    understanding.

Do not get me wrong, that's just a small hint for a feedback.

lib/mongo.ex
```diff
@@ -170,7 +170,12 @@ defmodule Mongo do
         query = query ++ [cursor: filter_nils(%{batchSize: opts[:batch_size]})]
         aggregation_cursor(conn, "$cmd", query, nil, opts)
       else
-        singly_cursor(conn, "$cmd", query, nil, opts)
+        if version >= 3 do
```
ankhers: We do not need to support versions less than 3.

lib/mongo.ex
```diff
@@ -790,6 +795,45 @@ defmodule Mongo do
...
+  """
+  @spec list_indexes(GenServer.server, String.t, Keyword.t) :: cursor
+  def list_indexes(topology_pid, coll, opts \\ []) do
+
+    with {:ok, conn, _, _} <- Mongo.select_server(topology_pid, :read, opts) do
+
+      c = aggregation_cursor(conn, "$cmd", [listIndexes: coll], nil, opts)
+
+      case Keyword.get(opts, :full, false) do
+        true -> c
```
ankhers: What is the reason for the `:full` option here? Do other drivers have a similar functionality?

lib/mongo.ex
```diff
@@ -790,6 +795,45 @@ defmodule Mongo do
...
+
+    ##
+    # from the specs
+    # https://github.com/mongodb/specifications/blob/f4bb783627e7ed5c4095c5554d35287956ef8970/source/enumerate-collections.rst#post-mongodb-280-rc3-versions
+    #
+    # In versions 2.8.0-rc3 and later, the listCollections command returns a cursor!
+    #
+    with {:ok, conn, _, _} <- Mongo.select_server(topology_pid, :read, opts) do
+      aggregation_cursor(conn, "$cmd", [listCollections: 1], nil, opts)
+      |> Stream.filter(fn coll -> coll["type"] == "collection" end)
```
ankhers, line 832: Is this actually necessary? I would think it should only return collections as is.

lib/mongo.ex
```diff
@@ -790,6 +795,45 @@ defmodule Mongo do
...
+  def show_collections(topology_pid, opts \\ []) do
+
+    ##
+    # from the specs
+    # https://github.com/mongodb/specifications/blob/f4bb783627e7ed5c4095c5554d35287956ef8970/source/enumerate-collections.rst#post-mongodb-280-rc3-versions
+    #
+    # In versions 2.8.0-rc3 and later, the listCollections command returns a cursor!
+    #
+    with {:ok, conn, _, _} <- Mongo.select_server(topology_pid, :read, opts) do
+      aggregation_cursor(conn, "$cmd", [listCollections: 1], nil, opts)
```
ankhers, line 831: We should be supporting all options that `listCollections` command accepts (https://docs.mongodb.com/manual/reference/command/listCollections/). Namely `filter` and `nameOnly`.

ankhers (commented):
How very tactful. The OS work I do is in my spare time. I do not get paid for it. When life gets busy, it is not my top priority. Please keep that in mind the next time you want to be "witty" with how you ask someone how much longer something will take.

I will go over the actual GridFS part of this when I have more time.

test/mongo/connection_test.exs
```diff
@@ -176,6 +176,9 @@ defmodule Mongo.ConnectionTest do
   end
 
   test "auth connection leak" do
+    # sometimes the function tcp_count() returns 1, so the test fails.
+    # maybe it is a good idea to wait a second before counting
+    :timer.sleep(1000)
```
ankhers, line 181: What if we just turn `async: false` at the top to prevent connections already being there? If that fixes it, maybe we could pull this single test out to it's own test module so that everything else can still run concurrenctly.

lib/mongo.ex
zookzook: removed support

lib/mongo.ex
zookzook: Ok, I split this into two functions. `list_index_names` is a convenient function to get only the names.

lib/mongo.ex
zookzook: You are right. I will change the code to support the options.

lib/mongo.ex
```diff
@@ -792,6 +793,45 @@ defmodule Mongo do
...
+    with {:ok, conn, _, _} <- Mongo.select_server(topology_pid, :read, opts) do
+      aggregation_cursor(conn, "$cmd", [listIndexes: coll], nil, opts)
+    end
+  end
+
+  @doc """
+  Convenient function that returns a cursor with the names of the indexes.
+  """
+  @spec list_index_names(GenServer.server, String.t, Keyword.t) :: %Stream{}
+  def list_index_names(topology_pid, coll, opts \\ []) do
```
ankhers, line 812: Do other drivers have this function, or something similar to it?

lib/mongo/grid_fs/bucket.ex
```diff
@@ -0,0 +1,229 @@
...
+
+  @spec delete(Bucket.t, BSON.ObjectId.t) :: Mongo.result(BSON.document)
+  def delete(%Bucket{topology_pid: topology_pid, opts: opts} = bucket, %BSON.ObjectId{} = oid) do
+    # first delete files document
+    collection = files_collection_name(bucket)
+    {:ok, %Mongo.DeleteResult{deleted_count: _}} = Mongo.delete_one(topology_pid, collection, %{_id: oid}, opts)
+
+    # then delete all chunk documents
+    collection = chunks_collection_name(bucket)
+    {:ok, %Mongo.DeleteResult{deleted_count: _}} = Mongo.delete_many(topology_pid, collection, %{files_id: oid}, opts)
```
ankhers, line 88: Use `with` here to be able to capture the errors instead of having match errors should something go wrong.

lib/mongo/grid_fs/bucket.ex
```diff
@@ -0,0 +1,229 @@
...
+  end
+
+  @doc """
+  Drops the files and chunks collections associated with
+  this bucket.
+  """
+  @spec drop(Bucket.t) :: Mongo.result(BSON.document)
+  def drop(%Bucket{topology_pid: topology_pid, opts: opts} = bucket) do
+    {:ok, _} = Mongo.command(topology_pid, %{drop: files_collection_name(bucket)}, opts)
+    {:ok, _} = Mongo.command(topology_pid, %{drop: chunks_collection_name(bucket)}, opts)
```
ankhers, line 98: Use `with`.

lib/mongo/grid_fs/bucket.ex
```diff
@@ -0,0 +1,229 @@
...
+    |> Enum.reduce(%Bucket{topology_pid: topology_pid, opts: options}, fn {k, v}, bucket -> Map.put(bucket, k, v) end)
+    |> check_indexes
+
+  end
+
+  @doc """
+  Returns the collection name for the files collection, default is fs.files.
+  """
+  @spec files_collection_name(Bucket.t) :: String.t
+  def files_collection_name(%Bucket{name: fs}), do: "#{fs}.files"
```
ankhers, line 52: Instead of `"#{fs}.files"`, could we use `fs <> ".files"`. It should be more performant. Same goes for all similar calls.

ankhers:
Thank you!

PR #242, opened 2018-10-25
adding x509 auth support

cadamis:
Some notes:
No tests, sorry. It'd probably take a fair bit for me to learn how to do that, but I can if needed.
I'm not entirely sure if adding the cipher to the `ssl_opts` is necessary for X509 by itself, but it's definitely necessary to connect to AWS, where I was testing this.
I didn't feel up to revising `Mongo.Auth.setup` to not require a password if connecting with X509, so for now, you need to provide a dummy string password in the `start_link` params, even though X509 doesn't require a password. The connection looks like this:
```
    {:ok, conn} = Mongo.start_link(
      database: "mydatabase",
      hostname: "mymongo.host.com",
      username: "LDAP valid username",
      password: "foo",  # still required unless Mongo.Auth.setup can be made smarter
      ssl: true,
      auth_mechanism: "MONGODB-X509",
      ssl_opts: [
        cacertfile: "/path/to/ca-cert.pem",
        certfile: "/path/to/cert.pem"
      ]
    )
```

### DIFF
--- a/lib/mongo/auth.ex
+++ b/lib/mongo/auth.ex
@@ -39,7 +39,7 @@ defmodule Mongo.Auth do
     if username && password, do: auth ++ [{username, password}], else: auth
   end
 
-  defp mechanism(%{wire_version: version, auth_mechanism: "MONGODB-X509"}) when version >= 3,
+  defp mechanism(%{wire_version: version, auth_mechanism: :x509}) when version >= 3,
     do: Mongo.Auth.X509
   defp mechanism(%{wire_version: version}) when version >= 3,
     do: Mongo.Auth.SCRAM
--- a/lib/mongo/auth.ex
+++ b/lib/mongo/auth.ex
@@ -39,6 +39,8 @@ defmodule Mongo.Auth do
     if username && password, do: auth ++ [{username, password}], else: auth
   end
 
+  defp mechanism(%{wire_version: version, auth_mechanism: "MONGODB-X509"}) when version >= 3,
+    do: Mongo.Auth.X509
   defp mechanism(%{wire_version: version}) when version >= 3,
     do: Mongo.Auth.SCRAM
   defp mechanism(_),
--- a/lib/mongo/protocol.ex
+++ b/lib/mongo/protocol.ex
@@ -90,7 +90,6 @@ defmodule Mongo.Protocol do
   defp ssl(%{socket: {:gen_tcp, sock}} = s, opts) do
     host      = (opts[:hostname] || "localhost") |> to_charlist
     ssl_opts = Keyword.put_new(opts[:ssl_opts] || [], :server_name_indication, host)
-    ssl_opts = Keyword.put_new(ssl_opts, :ciphers, ['AES256-GCM-SHA384'])
     case :ssl.connect(sock, ssl_opts, s.connect_timeout_ms) do
       {:ok, ssl_sock} ->
         {:ok, %{s | socket: {:ssl, ssl_sock}}}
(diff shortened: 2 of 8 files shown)


ankhers:
Thank you for this! I will take a look at it shortly.

lib/mongo/protocol.ex
```diff
@@ -89,6 +90,7 @@ defmodule Mongo.Protocol do
   defp ssl(%{socket: {:gen_tcp, sock}} = s, opts) do
     host      = (opts[:hostname] || "localhost") |> to_charlist
     ssl_opts = Keyword.put_new(opts[:ssl_opts] || [], :server_name_indication, host)
+    ssl_opts = Keyword.put_new(ssl_opts, :ciphers, ['AES256-GCM-SHA384'])
```
ankhers: This should not be the responsibility of the driver. The user can add this to their `:ssl_opts` if they need it.

lib/mongo/protocol/utils.ex
```diff
@@ -15,8 +15,14 @@ defmodule Mongo.Protocol.Utils do
...
   def command(id, command, s) do
-    op = op_query(coll: namespace("$cmd", s, nil), query: BSON.Encoder.document(command),
+    op = case command do
+      [authenticate: 1, user: _username, mechanism: "MONGODB-X509"] ->
+        op_query(coll: namespace("$cmd", nil, "$external"), query: BSON.Encoder.document(command),
                   select: "", num_skip: 0, num_return: 1, flags: [])
+      _command ->
+        op_query(coll: namespace("$cmd", s, nil), query: BSON.Encoder.document(command),
+                  select: "", num_skip: 0, num_return: 1, flags: [])
+    end
```
ankhers, line 23: This could probably be cleaned up to remove the duplication. To make this less brittle, you may want to use `Keyword.get/2` to grab only the important parts of the X509 message.

lib/mongo/auth.ex
```diff
@@ -39,6 +39,8 @@ defmodule Mongo.Auth do
     if username && password, do: auth ++ [{username, password}], else: auth
   end
 
+  defp mechanism(%{wire_version: version, auth_mechanism: "MONGODB-X509"}) when version >= 3,
```
ankhers: I would rather the `:auth_mechanism` option to be an atom instead of a string.

cadamis: The JS and Python libs use a string, and the official docs kind of point that way too. (Check the command line client connection at the bottom of https://docs.mongodb.com/manual/tutorial/configure-x509-client-authentication/). Java uses an x509 connection subclass. But, you're the author, so fine by me.

lib/mongo/protocol.ex
cadamis: I went back and forth on this, because you're right, but also digging out this arcane tidbit about Erlang's SSL from the depths of the internet was a huge pain. I think a good in-between will be to create an `examples` dir where people can contribute specific connection params that might be non-obvious. I'll add one to my PR. Hopefully this works for you.

lib/mongo/protocol/utils.ex
cadamis: I can definitely extract out the namespace difference to reduce duplication. I'll try setting the namespace with `Keyword.get/2` but I'm not sure if it's less brittle. I think the case looks a little better, but I'll update the PR and you can decide.

ankhers:
I think this is good. Can you just write a quick piece of documentation on `Mongo.start_link/1` about the new `:auth_mechanism` option?

cadamis:
Readme updated with auth mechanism info.

ankhers:
Sorry, I meant part of the [function doc](https://github.com/ankhers/mongodb/blob/master/lib/mongo.ex#L82-L116).

cadamis:
Yeah actually that makes more sense. Added.

ankhers:
Thank you!

PR #251, opened 2018-11-21
Upgrade dialyzer rc4

gpad:
- Upgrade dialyzer to rc 4 
- Add new flags to dialyzer underspecs, unknown, unmatched_returns
- Fixed the new warning
- Fix some spec and add some tests


lib/mongo.ex
```diff
@@ -61,7 +61,7 @@ defmodule Mongo do
   @type conn :: DbConnection.Conn
   @type collection :: String.t
   @opaque cursor :: Mongo.Cursor.t | Mongo.AggregationCursor.t | Mongo.SinglyCursor.t
-  @type result(t) :: :ok | {:ok, t} | {:error, Mongo.Error.t}
+  @type result(t) :: :ok | {:ok, t} | {:ok, nil}| {:error, Mongo.Error.t}
```
ankhers, line 64: Is this `{:ok, nil}` only there because of `direct_command/3`? If so, I think we should look whether or not that part is actually needed. I would rather avoid returning `nil` if possible, I'm guessing most people are not checking for that.

gpad: If you check the test file `test/mongo_test.exs` 

```
    assert {:ok, nil} == Mongo.find_one_and_update(c.pid, coll,
      %{"number" => 666},
      %{"$set" => %{title: "the number of the beast"}})
```
you can see that it returns `{:ok, nil}` also if it doesn't find anything.

ankhers: Could we instead change the spec on `find_one_and_update/4` and `direct_command/3` to say that it could return a `nil`? I just don't want to give the impression that everything can return `nil`.

ankhers:
Thanks!

PR #254, opened 2018-12-03
Allow passing read_preference without all keys

sauvainr:
Currently specify the ```read_preference``` requires to pass all keys as done in 
https://github.com/ankhers/mongodb/blob/65bd68adbc2753c38d927ad0ba804fe8e66d50d7/test/mongo/topology_description_test.exs#L10

This is confusing and not handy. I made this commit to support it.


ankhers:
Thank you!

PR #256, opened 2018-12-04
fix parse write concern from url

### LINKED ISSUE #255: Can't set numeric value for write concern in url
When connection string has write concern set to numeric value
`mongodb://mongo-1,mongo-2,mongo-3/db-name?w=2`
write operations fail with following message:
`No write concern mode named '2' found in replica set configuration`
Look like url parser does not recognize numeric values.

rblack:
resolve #255


ankhers:
Thank you!

PR #259, opened 2018-12-06
Respect secondary_preferred read preference by falling back to primary

tjsousa:
As this read preference is following the exact same code path as the secondary
read preference, this has the side-effect of not being able to make read
operations when only the primary is available.

According to the [MongoDB documentation](https://docs.mongodb.com/manual/core/read-preference-mechanics/#replica-set-read-preference-behavior) for the `secondary_preferred`:
>  1. Following the server selection process for the read preference secondary, if a list of eligible secondary members is non-empty, driver chooses an eligible secondary member.
> 2. Otherwise, if the list is empty, driver selects the primary.

This change ensures this fallback and simplifies the primary_preferred
implementation to mimic the behaviour, only the other way around.

### DIFF
--- a/lib/mongo/topology_description.ex
+++ b/lib/mongo/topology_description.ex
@@ -92,32 +92,39 @@ defmodule Mongo.TopologyDescription do
 
   ## Private Functions
 
-  defp select_replica_set_server(topology, mode, read_preference)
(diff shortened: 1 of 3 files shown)


ankhers:
Thank you!

Issue #250, opened 2018-11-15
Read preference for secondary nodes has strange incosistent behaviour with different operations


sashaafm:
Version: 0.4.7
Elixir: 1.5.2

We're working on a project which we want to read only from secondary nodes of our Mongo cluster. To achieve this we set our options initially as so:
```
options = [
      sort: %{"_id" => 1}, # Sorts in ascending order
      skip: (page - 1) * per_page,
      pool: DBConnection.Poolboy,
      read_preference: Mongo.ReadPreference.defaults(%{mode: :secondary_preferred}),
     slave_ok: true
    ]
```
And we perform two different read operations:

1. Find All: this operation uses a given field to return all the objects with that field and how many there are. First, it uses `Mongo.count!/4` to get how many objects exist with that field, followed by a `Mongo.find/4`, which retrieves all those objects fitting into a given page size.
2. Find By ID: this operation return an object with a given field (like the previous op) and a given ID. It only relies on `Mongo.find/4` to fetch that object.

With the above configuration both operations work, but we weren't able to determine if the driver is in fact reading from the secondary nodes.

However, a strange behaviour arises when setting `slave_ok: false`. With this option set as `false`, the second operation works, but the first fails with 
```
 message: "command failed: not master"}, stack: [{Mongo, :count!, 4, [file: 'lib/mongo.ex', line: 311]}
```

This feels very strange to us. As specified here in the Mongo [Read Preference docs](https://docs.mongodb.com/manual/core/read-preference/#read-preference-modes) we're supposed to get this behaviour:

> In most situations, operations read from secondary members but if no secondary members are available, operations read from the primary.

Since we've got both primary and secondary nodes up in our cluster, we should be reading from secondary nodes.

Continuing on and reading about [slaveOk](https://docs.mongodb.com/manual/reference/method/rs.slaveOk/index.html#rs-slaveok) we understand that this allows the connection to read from secondary nodes.

So, let's summarise this:

1. Read Preference to secondary tells the driver to read from the secondary nodes whenever possible.
2. `slaveOk` set to `true` actually allows the driver to read from the secondary nodes.

Then we've got these two configurations as explained above:

| Configuration |   Read Preference   | SlaveOk |                           Result                           |
|:----------:|:-------------------:|:-------:|:----------------------------------------------------------:|
|      1     | secondary_preferred |   true  | Both operations work, but unsure if reading from secondary |
[...]

PR #261, opened 2018-12-11
Review usage of `slave_ok` flag according to the driver spec

tjsousa:
This PR directly addresses the inconsistencies uncovered in #250, where the usage of a read preference allowing reads from secondaries would not play well with the value of `slave_ok`, which in the driver defaults to `false` (not setting the flag).

It follows directions in https://github.com/mongodb/specifications/blob/master/source/server-selection/server-selection.rst#use-of-slaveok and tackles the following issues:
1. Inconsistencies regarding `slave_ok`'s implicit behavior
2. `findAndModify` operations relying on a `read` server selection (instead of `write`)

Moreover, for cursor-based queries the `slave_ok` option is already implicitly set (and reused for further `getmore` invocations) but commands such as `count` would fail with a non-primary read preference (because `slave_ok` would be set to false). 

The way to workaround this limitation would be to use the deprecated and undocumented behaviour of passing `slave_ok: true` as a query option, so that it would override the default and allow some commands to go through on secondaries. A side-effect of this would be all the `findAndModify` operations (`find_one_and_update`, `find_one_and_replace` and `find_one_and_delete`) to also go through secondaries, which the default `slave_ok` prevented, but should never be possible (as these are equivalent to write operations).

The driver specification goes even further and mentions this combination of options used for the workaround as an error:
> If a driver API allows users to potentially set both the legacy slaveOK configuration option and a default read preference configuration option, passing a value for both MUST be an error. (See Use of slaveOk for the two uses of slaveOK.)

This PR removes the need for the workaround by having `slave_ok` consistent with the server selection algorithm but doesn't prohibit the combination of options as the backwards compatibility for this option is not totally clear.

### DIFF
--- a/lib/mongo.ex
+++ b/lib/mongo.ex
@@ -212,7 +212,7 @@ defmodule Mongo do
 
     opts = Keyword.drop(opts, ~w(bypass_document_validation max_time projection return_document sort upsert collation)a)
 
-    with {:ok, conn, _, _} <- select_server(topology_pid, :read, opts),
+    with {:ok, conn, _, _} <- select_server(topology_pid, :write, opts),
          {:ok, doc} <- direct_command(conn, query, opts), do: {:ok, doc["value"]}
   end
 
@@ -252,7 +252,7 @@ defmodule Mongo do
 
     opts = Keyword.drop(opts, ~w(bypass_document_validation max_time projection return_document sort upsert collation)a)
 
-    with {:ok, conn, _, _} <- select_server(topology_pid, :read, opts),
+    with {:ok, conn, _, _} <- select_server(topology_pid, :write, opts),
          {:ok, doc} <- direct_command(conn, query, opts), do: {:ok, doc["value"]}
   end
 
@@ -283,7 +283,7 @@ defmodule Mongo do
     ] |> filter_nils
     opts = Keyword.drop(opts, ~w(max_time projection sort collation)a)
 
-    with {:ok, conn, _, _} <- select_server(topology_pid, :read, opts),
+    with {:ok, conn, _, _} <- select_server(topology_pid, :write, opts),
          {:ok, doc} <- direct_command(conn, query, opts), do: {:ok, doc["value"]}
   end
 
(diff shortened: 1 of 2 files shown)


lib/mongo.ex
```diff
@@ -515,7 +514,8 @@ defmodule Mongo do
   def command(topology_pid, query, opts \\ []) do
     rp = ReadPreference.defaults(%{mode: :primary})
     rp_opts = [read_preference: Keyword.get(opts, :read_preference, rp)]
-    with {:ok, conn, _, _} <- select_server(topology_pid, :read, rp_opts),
+    with {:ok, conn, slave_ok, _} <- select_server(topology_pid, :read, rp_opts),
+         opts = Keyword.put(opts, :slave_ok, slave_ok),
```
mvpgomes, line 518: 👍

lib/mongo.ex
```diff
@@ -212,7 +212,7 @@ defmodule Mongo do
 
     opts = Keyword.drop(opts, ~w(bypass_document_validation max_time projection return_document sort upsert collation)a)
 
-    with {:ok, conn, _, _} <- select_server(topology_pid, :read, opts),
+    with {:ok, conn, _, _} <- select_server(topology_pid, :write, opts),
```
mvpgomes, line 215: 👍

lib/mongo/topology_description.ex
```diff
@@ -92,7 +92,7 @@ defmodule Mongo.TopologyDescription do
 
   ## Private Functions
 
-  defp select_replica_set_server(topology, :primary, read_preference) do
+  defp select_replica_set_server(topology, :primary, _) do
```
mvpgomes: 💅 ❤️

mvpgomes (approved):
Awesome work here @tjsousa 👏

lib/mongo/topology_description.ex
ankhers: Since the `read_preference` variable is unused, can you just call it `_read_preference` instead?

ankhers:
Thanks for this. If you can make that quick change, I will merge this.

lib/mongo/topology_description.ex
tjsousa: Will do.

tjsousa:
@ankhers Pushed the change you've requested. Thanks for looking at this.

ankhers:
Thanks again!

PR #266, opened 2019-01-08
clarifying the README for poolboy usage

cadamis:
Hopefully addresses a couple (theoretically) common question about `:pool` usage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,7 +100,13 @@ def start(_type, _args) do
 end
 ```
 
-Then you can use the pool as following:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -100,7 +100,13 @@ def start(_type, _args) do
 end
 ```
 
-Then you can use the pool as following:
+DBConnection defaults to [10 Poolboy connections](https://hexdocs.pm/db_connection/1.1.3/DBConnection.Poolboy.html#content), but you can change that with the `:pool_size` option:
```
ankhers: Can you change `DBConnection` to `DBConnection.Poolboy` to prevent people from thinking that this is the default behaviour? It should be clear that this is only when using the poolboy specific module.

cadamis: Oh, I didn't realize that, good clarification.

ankhers:
Thanks!

PR #270, opened 2019-01-29
Fixes time units atoms deprecation warnings

pchaussalet:
Plural time units are deprecated since Elixir 1.8.0 as singular ones are supported since v1.4.0.


ankhers:
Thank you!

PR #283, opened 2019-05-09
Add missing host parameter to exception call

tschmittni:
send_error and recv_error create a Mongo.Error without
host and this fails with the following error message:

no function clause matching in Mongo.Error.exception/1
Mongo.Error.exception([tag: :tcp, action: \"recv\", reason: :timeout])


ankhers:
Thank you!

PR #286, opened 2019-05-14
Update pool documentation.

ColaCheng:
`DBConnection.Poolboy` is no longer available in DBConnection 2.0.6. I created the PR to correct the pool documentation part.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ Then run `mix deps.get` to fetch dependencies.
 
 ### Connection pooling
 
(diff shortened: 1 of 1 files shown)


ankhers:
I don't think the changes you made are strictly correct. I believe db_connection only ships with a single pool now. So there is no need to pass in the `:pool` option at all. It will just do the right thing based on the `:pool_size` option.

Would you mind making those changes, or would you like me to?

ColaCheng:
Thank you for your notice. I didn't know this part. I updated the documentation. By the way, we don't need to specify the pool in each query right?

ankhers:
No, since there is only the one pool now, we should not need to specify the `:pool` in every query.

ankhers:
Thank you!

PR #289, opened 2019-05-21
stop old connection pool

Sanchos01:
if application and db on different hosts and mongo restarted, app creates more than one connection pools


ankhers:
Would it be possible to try and build a test that this passes?

Sanchos01:
Test added which one simulate reconnection

ankhers:
Thank you!

PR #291, opened 2019-05-22
Fix warnings and remove SinglyCursor module

gergo-papp:
- I've seen these warnings a thousand times. I thought it's time to fix them...
- I also removed some unused functions in lib/mongo.ex.


ankhers:
Thank you for this! Since we no longer use `singly_cursor/5`, would you mind removing the `Mongo.SinglyCursor` module, as well as any references to it?

gergo-papp:
> Thank you for this! Since we no longer use `singly_cursor/5`, would you mind removing the `Mongo.SinglyCursor` module, as well as any references to it?

Sure!

ankhers:
Thank you!

Issue #290, opened 2019-05-21
BSON.Binary decoding issue for :binary_old subtype


skinandbones:
I'm having an issue using Mongo.GridFS.Download where the output files are 4 bytes longer than I expect them to be.  The data was inserted with the ruby driver around 2013 using some of the early early versions of things.  These old documents can be decoded successfully using the current up-to-date ruby mongo drivers.

Note: I do not having problems decoding more recent data.

I've debugged up to the point where I've noticed the binary data in the .chunks document is of subtype `:binary_old`.   These are the ones where I'm having trouble.   If I remove the first 4 bytes from the binary data, then I get the expected results.

I'm hoping for any quick ideas on the issue.   I will see if I can create a failing test or devise a patch.  Unfortunately, the BSON spec and binary strings in Elixir aren't my specialty :)

ankhers:
Unfortunately I'm not sure off the top of my head. If you can create a failing test, I may be able to help a little bit more.

skinandbones:
I have a fix but it's going to need a little ❤️before PR.

According to http://bsonspec.org/spec.html, for the BinData datatype if the subtype is 0x02 (OLD), then there's an extra int32 in the binary data which is the length of the bytes.  A fix looks like this,

```
diff --git a/lib/bson/decoder.ex b/lib/bson/decoder.ex
index 56cbf54..9456282 100644
--- a/lib/bson/decoder.ex
+++ b/lib/bson/decoder.ex
@@ -51,6 +51,11 @@ defmodule BSON.Decoder do
     list(binary)
   end

+  defp type(@type_binary, <<_size::int32, subtype, length::int32, binary::binary(length), rest::binary>>) when subtype == 0x02 do
+    subtype = subtype(subtype)
+    {%BSON.Binary{binary: binary, subtype: subtype}, rest}
+  end
+
   defp type(@type_binary, <<size::int32, subtype, binary::binary(size), rest::binary>>) do
     subtype = subtype(subtype)
     {%BSON.Binary{binary: binary, subtype: subtype}, rest}
```

That seems to do the trick.   I got a clue by reading the bson-ruby code:
https://github.com/mongodb/bson-ruby/blob/master/lib/bson/binary.rb#L155

ankhers:
Resolved in #292

PR #292, opened 2019-05-22
Add support for the :binary_old subtype for BSON.Binary

skinandbones:
The problem is discussed in Issue #290.

This PR is a work in progress.  The proposed code change is working on my production dataset. 
Remaining is devising a test strategy.  I don't think it's going to be possible to insert the old binary format into a test DB so we may need to do something else.


skinandbones:
Tests added according to my understanding of the BSON spec for binary_old.

ankhers:
Thank you!

PR #296, opened 2019-06-13
Add LDAP Auth for MongoDB Enterprise

churchie317:
Builds upon the 0.4.7 release.

I know this isn't a burning need in the community, but it would be super helpful if we could push this into a 0.4.8 release.

This change would extend current functionality by allowing for LDAP Authorization via passing the `auth_mechanism: :plain` option to the start_link function.

Thanks!

P.S. I don't expect this to be mergeable into the origin `master` branch. Could we create a new branch from the 0.4.7 release and merge into that?


churchie317:
Also, I'm happy to implement these changes in the > 0.5.1 releases if the community would find that helpful.

ankhers:
I think the easiest thing to do would be to write this PR for the current master and then I should be able to backport the commit and do a 0.4.8 release. 

Though, if you can, please avoid running `mix format` on the files. I know it is something that needs to be done, but it makes it harder to find what you actually changed in your PR. I would rather just run it for the entire project in one go.

Issue #273, opened 2019-03-05
Can't connect to AWS DocumentDB


stewartdennis:

ankhers:
This doesn't give me any information I can use to debug the issue. Are you getting an error message or anything? Are you willing to give me access to a database to look into the issue?

ninad1487:
Hi @ankhers,

We're trying to connect AWS DocumentDB from Elixir/Phoenix framework.

Seems like the authentication has an issue as below lines:
https://github.com/ankhers/mongodb/blob/master/lib/mongo/auth/scram.ex#L33
https://github.com/ankhers/mongodb/blob/master/lib/mongo/auth/scram.ex#L54

The function expects `conversationId : 1`, but we are observing different values than 1. After fixing this locally, any query returns the error `errmsg: Legacy opcodes are not supported, code: 303`. It seems that documentDB supports mongodb 3.6 API, but the driver is based on older opcodes. We would appreciate any thoughts on this.

We’d be open to providing access to a test database and environment if needed.

stewartdennis:
@ankhers Please see https://github.com/AlimovSV/mongodb/tree/docdb for a temporary fix.

ankhers:
Thank you for following up. I believe I have actually started work on moving away from the older opcodes but never finished the work. But I would be glad to finish the work I had previously started. If you could contact me privately (you should be able to find my email on my profile page here) in order to get access to a test database that I can verify my changes against, it would be appreciated.

ankhers:
Closing due to inactivity. If you get me the information I requested, I would be glad to reopen this.

christopherlai:
@ankhers I ran into a similar issue as well.

`Mongo.Auth.SCRAM.first/5` expects `conversationId` to be `1`, but in my situation that number was much higher (`17842` as an example).

I would be open to setting up a DocumentDB cluster for troubleshooting this issue.

christopherlai:
I also noticed that `Mongo.Auth.SCRAM.second/2` receives `"done" => true` instead of `false`.

jethro-solve:
We've run into the same issue when testing feasibility of using DocumentDB with an Elixir application. I've taken the work referenced above by @AlimovSV and refactored it so that it's mergable against master again for easy testing/experimenting.

https://github.com/ankhers/mongodb/compare/master...solvedata:master

With these changes, I can successfully authenticate to DocumentDB with TLS disabled (haven't looked into TLS yet) and run queries.

[...]

PR #304, opened 2019-07-17
Add support for AWS DocumentDB

jethro-solve:
As per thread in #273, there are some changes required to add support for DocumentDB. These were written by @AlimovSV originally, I've just re-factored them to get a mergeable PR.

Note that I'm not properly across what is going on in this code, so it warrants careful attention and testing.

### DIFF
--- a/lib/mongo.ex
+++ b/lib/mongo.ex
@@ -555,13 +555,6 @@ defmodule Mongo do
            direct_command(conn, query, opts) do
       {:ok, %{from: 0, num: Enum.count(docs), cursor_id: id, docs: docs}}
     end
-
-    # params = [query, select]
-    # query = %Query{action: :find, extra: coll}
-    # with {:ok, reply} <- DBConnection.execute(conn, query, params, defaults(opts)),
-    #      :ok <- maybe_failure(reply),
-    #      op_reply(docs: docs, cursor_id: cursor_id, from: from, num: num) = reply,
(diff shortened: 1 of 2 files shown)


foggy1:
@jethro-solve I may make a PR if I manage to track down what's happening (I literally just discovered this and am happy it had an explanation!) but this PR will break the work `cursor.ex` is doing because in line 77 of that file when it runs `Mongo.get_more` the new returned value appears to match what should be returned by the _first_ batch but not what should be returned from the "next" batches.

I'm going to get more specific shortly but just wanted to shout about it because I'm excited.

foggy1:
To be clear, `Mongo.Cursor.next_fun/2` has a pattern match that goes as such:

```
case Mongo.get_more(conn, coll, cursor, opts) do
            {:ok, %{"cursor" => %{"id" => cursor, "nextBatch" => []}}} ->
              {:halt, state(state, cursor: cursor)}

            {:ok, %{"cursor" => %{"id" => cursor, "nextBatch" => docs}}} ->
              {docs, state(state, cursor: cursor, limit: new_limit(limit, batch_size))}

            {:error, error} ->
              raise error
          end
```

But the proposed changes make it so that `Mongo.get_more` returns something looking like:
```
{:ok, %{from: 0, num: Enum.count(docs), cursor_id: id, docs: docs}}
```

So you get a `no case clause matching` error on anything that needs more than reduces more than the first response from the cursor.

PR #309, opened 2019-08-07
Use internal call to select_server

hauleth:
This allows to make `select_server` function private in future release.


ankhers:
I'm not really sure what is going on with Travis right now with Erlang 18, but all of the other tests appear to be passing, so I will merge this. Thank you!

PR #317, opened 2019-10-02
Implement Decimal128.decode

tschmittni:
Converting Decimal into binary so that it can actually be stored.
The PR https://github.com/ankhers/mongodb/pull/221 only added
support for querying but not inserting decimals.


ankhers:
Thank you!

PR #318, opened 2019-10-04
Add Topology.has_readable_server? real implementation

esse:
This PR adds `has_readable_server?` method to `TopologyDescription`.

The previous implementation returned always `true` which wasn't right in many cases.


ankhers:
Thank you!

PR #339, opened 2020-06-02
Add Kobil as license holder

hauleth:
Additionally it cleanups README a little

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,23 +1,24 @@
-# Mongodb
+# MongoDB
 
 [![Build Status](https://travis-ci.org/ankhers/mongodb.svg?branch=master)](https://travis-ci.org/ankhers/mongodb)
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -29,38 +30,42 @@
...
+when it's decoded. Additionally the driver will accept both atoms and strings
+for document keys but will only decode to strings.
+
+² BSON symbols can only be decoded.
 
 ## Usage
 
 ### Installation:
 
-Add mongodb to your mix.exs `deps` . Mongodb supports the same pooling libraries db_connection does (currently: no pooling, poolboy, and sbroker). If you want to use poolboy as pooling library you should set up your project like this:
```
hauleth, line 57: This is removed as it is no longer true as we currently use `db_connection` 2.x which has custom pool implementation and do not support neither Poolboy nor Sbroker.

PR #341, opened 2020-06-25
Custom Mongo.Encoder protocol

esse:
Hi,

This PR adds a feature: `Mongo.Encoder` protocol.

In our project, we very often spot a case when we need to insert map, generated from some struct into the mongo database. With that PR, instead of writing a custom glue code, we can simply implement `Mongo.Encoder` protocol. Which in turns results in clearer, more idiomatic Elixir code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,48 @@ for document keys but will only decode to strings.
 
 ² BSON symbols can only be decoded.
 
+### Writing your own encoding info
+
+If you want to write a custom struct to your mongo collection - you can do that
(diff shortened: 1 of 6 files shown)


deadtrickster:
I like the idea. How you do decoding back to domain structs?

deadtrickster (approved):

hauleth (approved):

esse:
@deadtrickster thanks! Right now, we're decoding them manually - and I believe there is no such elixir idiomatic way of doing that, such as here.

deadtrickster:
yeah I think it is ecto territory.  we should look into it.

PR #351, opened 2020-11-06
test fix and compiler warning eliminations

kbaird:


test/mongo/config_hide_test.exs
```diff
@@ -2,8 +2,6 @@ defmodule Mongo.HideConfigTest do
   use ExUnit.Case, async: true
   alias Mongo.ConfigHide
 
-  @password_masked "***"
```
kbaird, line 5: This was unsed

test/mongo/url_parser_test.exs
```diff
@@ -15,7 +15,7 @@ defmodule Mongo.UrlParserTest do
 
       opts = UrlParser.parse_url(url: url)
       password = Keyword.get(opts, :password)
-      assert password = "2yPK}Bzj|qE($^1JDdk4J42*&4lLgV%C"
+      assert password == "2yPK}Bzj|qE($^1JDdk4J42*&4lLgV%C"
```
kbaird, line 18: I presume this is what you actually want here? I'm not sure what a pattern match would accomplish.

hauleth: In this case these two are equivalent.

kbaird: I think
```elixir
assert ^password = "2yPK}Bzj|qE($^1JDdk4J42*&4lLgV%C"
```

would be equally useful, but what's there now would just re-assign the value of `password` if it differs.

kbaird: The existing code just changes the value in `password` while still passing `assert`, because it's truthy:
```
iex(1)> password = "old password"
"old password"
iex(2)> password == "2yPK}Bzj|qE($^1JDdk4J42*&4lLgV%C"
false
iex(3)> ^password = "2yPK}Bzj|qE($^1JDdk4J42*&4lLgV%C"
** (MatchError) no match of right hand side value: "2yPK}Bzj|qE($^1JDdk4J42*&4lLgV%C"

iex(3)> force_into_boolean = !!(password = "2yPK}Bzj|qE($^1JDdk4J42*&4lLgV%C")
true
iex(4)> password
"2yPK}Bzj|qE($^1JDdk4J42*&4lLgV%C"
```

scottmessinger:
I think these were cleaned up so closing this. Thanks for submitting it and I'm sorry it was never merged!!!

PR #356, opened 2021-04-09
Replace hmac/3 calls with mac/4 calls

romynior:
hmac/3 is now obsolete and the recommended function is mac/4


danyflorez:
Is this gonna be fixed someday? Supporting OTP 24 would be great

PR #357, opened 2021-04-09
Replace hmac/3 calls with mac/4 calls (#1)

romynior:
hmac/3 is now obsolete and the recommended function is mac/4


lenaw:
Hi, is it possible to merge this PR? because after upgrading to OTP 24, mongo is crashing because `:crypto.hmac/3` does not exist anymore.

danyflorez:
Hi, can we merge this PR? we really need to update to OTP 24

PR #359, opened 2021-04-20
Bug 1402675 add support for tls in url parser

lnowotny:
Mongo database connections could not be established when 'tls=true' was included in connection strings. With this change, the presence of 'tls=true' is the same as 'ssl=true'.


lnowotny:
Inadvertently picked up extra commit

PR #360, opened 2021-06-28
change hmac to mac

jrzayev:


IgorPolyakov:
Hello!) 
If you don't want to break backward compatibility
you can use some thing like this 

```elixir
 if System.otp_release() >= "22" do
    defp hmac_fun(digest, key), do: &:crypto.mac(:hmac, digest, key, &1)
  else
    defp hmac_fun(digest, key), do: &:crypto.hmac(digest, key, &1)
  end
```
source https://github.com/elixir-plug/plug_crypto/pull/20/files

scottmessinger:
@IgorPolyakov that's a great suggestion. @Javid907 Could you make that change? That would be very helpful!

joeapearson:
Addressed in #364 although without backwards compatibility.  If backwards compatibility is desired please can we make that a new PR?  Thanks for bearing with us while we clean up these issues.

PR #362, opened 2021-08-20
Fix moduledoc of UpdateResult :upserted_ids

1st8:
It was formerly `:upserted_id` and is now `:upserted_ids`, but it seems the `@moduledoc` was not yet updated.


joeapearson:
Looks good.  Thank you.

PR #364, opened 2021-09-16
Ecto 3 supporting work

joeapearson:
Supporting changes for Ecto 3 work.

New APIs:

* update/5 - directly address Mongo update command which allows for more complex update queries.
* Add support for OTP 24.

Incompatibilities:

* OTP versions earlier than 24 not supported.  This could be addressed by adding some backwards compatibility code.
* `find_one_and_update`, `find_one_and_replace`, `find_one_and_delete` operations now return a `%Mongo.FindAndModifyResult` struct.  Existing clients that depend on this should refer to the structs `value` field.  This was required in order to implement upsert functionality in `mongodb_ecto` which needs the extra information returned by mongo in order to identify insert vs update operations.

If these changes are accepted I suggest we publish them under a `-beta` version in Hex.  

I want to add github actions to this repo for ongoing verification but I suggest that is a separate PR.


scottmessinger:
@joeapearson I'm in favor of a beta tag! Also, this library has never had a 1.0 release, so I think it's fine to publish this as 1.0-beta and then cut the full 1.0 in a month or so if everything is looking good.

sa-mm:
I’m curious what part of this PR requires OTP 24? I see the commit for the new crypto api, but that is new as of 22.1 according to the docs (https://erlang.org/doc/man/crypto.html#mac-3).

joeapearson:
> I’m curious what part of this PR requires OTP 24? I see the commit for the new crypto api, but that is new as of 22.1 according to the docs (https://erlang.org/doc/man/crypto.html#mac-3).

Good spot.  It looks like you're right and perhaps we secretly do have support for OTP 22.1.  This remains completely untested (by me at least), I suggest adding some kind of CI test in #365 to verify.

PR #370, opened 2021-09-27
GitHub actions

joeapearson:
- [x] Matrix github actions
- [x] Add credo for linting
- [x] Mix format
- [x] Add docker compose for mongo, including replicaset support

I've added support for matrix builds.  Currently there is only one combination:

1. Ubuntu latest
2. MongoDB 4.4
2. Elixir 1.12.3
3. OTP 24.0.6

We can add more combinations later.


scottmessinger:
I'm really excited about this!

For our app, we've been using this to start mongodb: 

```

      - name: Start MongoDB
        uses: supercharge/mongodb-github-action@1.3.0
        with:
          mongodb-version: 4.0.27
          mongodb-replica-set: "replica-set"
```

I'm wondering if this might help solve the problem where the tests are failing b/c they can't connect to Mongodb.

joeapearson:
> I'm really excited about this!

Thanks I am too!  I think it's going to help us a lot to accept changes going forward.

> I'm wondering if this might help solve the problem where the tests are failing b/c they can't connect to Mongodb.

That might be the way we have to go.

For now I'm trying to go down the docker-compose approach because then there is a guaranteed identical environment in dev and CI.

I've nearly got it running on my machine (most of the tests pass too!).

It looks like:

1. SSL tests are always disabled in test.  Let's fix this one day.  I propose not to fix this in this branch and create a separate issue.
2. Docker test are always disabled in test.  Similarly, I think this could be fixed in a separate issue but is less important IMO.

I'd like to report test coverage out too but that'll be a separate effort.

joeapearson:
> For our app, we've been using this to start mongodb:

Scott FYI I tried this and it did get further but since the supercharge action only creates a single member replicaset some of the tests that exercise different write concerns won't work.

joeapearson:
@scottmessinger would you mind reviewing this please.  There are a number of outstanding issues I suggest we tackle in separately:

1. SSL tests are not exercised during CI (this has always been the case, they are disabled during CI)
2. Socket tests are not exercised during CI (ditto)
3. Test coverage is not reported as part of the process
4. We only run tests with a single combination of Mongo 4.4, Elixir 1.12 etc.
5. One test fails and I'm not sure why!
6. The github action manually installs and starts Mongo at the specified version.  The code is ugly.  This would be fine except github actions appears to cache the ubuntu image between builds, such that we aren't guaranteed a clean build every time.  My preference is to use docker-compose to solve this but...
6. I abandoned my docker-compose work for now because I just don't have time to make it happen yet.  The tests (see `test_helper.exs` expects to be able to use the `mongo` CLI directly in order to perform some manual setup.  I don't like this but it is tricky to work around.
7. The code-quality (particularly credo) has a bunch of issues that ought to be fixed.  Not a huge issue but the consistency makes it easier for folks to work on.

If you're happy I suggest you merge this in, despite it not being perfect (unless you disagree of course!).  We'll have some validation going forward and we can pick these issues off one by one.  Then I'll create an issue for each of the points above; some of them are excellent starter issues so we can take advantage of some of the recent enthusiasm for contributions we've had too.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,80 @@
+on: push
```
ankhers: I would suggest this being

```yaml
on:
  pull_request:
  push:
    branches:
      - master
```

This way the CI is only run for pull requests and when pushing to master.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,80 @@
...
+          # These can be upgraded as and when but probably won't need to change much.
+          otp-version: "24.0.6"
+          elixir-version: "1.12.3"
+      - uses: actions/cache@v2
+        with:
+          path: |
+            deps
+            _build
+            priv/plts
+          key: ${{ runner.os }}-mix-${{ hashFiles('**/mix.lock') }}
```
ankhers, line 24: It may be worth including the Elixir and OTP versions in the key as well. It may not be super important for code quality, seeing as you will probably only care about the most recent version of elixir.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,80 @@
...
+      - run: mix dialyzer
+  test:
+    strategy:
+      matrix:
+        os: 
+          - ubuntu-latest
+        otp-version:
+          - "24.0.6"
+        elixir-version: 
+          - "1.12.3"
```
ankhers: We should probably setup an actual matrix  and test previous versions of Elixir and OTP as well.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,80 @@
...
+      - uses: erlef/setup-beam@v1
+        with:
+          otp-version: ${{ matrix.otp-version }}
+          elixir-version: ${{ matrix.elixir-version }}
+      - uses: actions/cache@v2
+        with:
+          path: |
+            deps
+            _build
+          key: ${{ runner.os }}-mix-${{ hashFiles('**/mix.lock') }}
```
ankhers: If there is a proper matrix of Elixir and OTP versions, you will want to include the versions in the cache key in order to pull the proper cache version.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,80 @@
...
+          sudo apt-get update
+
+          # Finally install mongodb
+          sudo apt-get install -y \
+            --allow-downgrades \
+            mongodb-org=${{ matrix.mongo-version.maj }}.${{ matrix.mongo-version.min }}.${{ matrix.mongo-version.patch }} \
+            mongodb-org-server=${{ matrix.mongo-version.maj }}.${{ matrix.mongo-version.min }}.${{ matrix.mongo-version.patch }} \
+            mongodb-org-shell=${{ matrix.mongo-version.maj }}.${{ matrix.mongo-version.min }}.${{ matrix.mongo-version.patch }} \
+            mongodb-org-mongos=${{ matrix.mongo-version.maj }}.${{ matrix.mongo-version.min }}.${{ matrix.mongo-version.patch }} \
+            mongodb-org-tools=${{ matrix.mongo-version.maj }}.${{ matrix.mongo-version.min }}.${{ matrix.mongo-version.patch }}
```
ankhers: Maybe instead of this, you can look at what [one of the official mongo drivers](https://github.com/mongodb/mongo-ruby-driver/blob/master/.github/workflows/test.yml) uses to start mongo.

CONTRIBUTING.md
```diff
@@ -0,0 +1,16 @@
+# Contributing
+
+Thanks for considering making a contribution to `mongodb`!
+
+## Getting started
+
+1. Clone this repository
+2. Install mongodb and make sure it's available on `$PATH`
+3. Run `start_mongo.bash`
```
ankhers, line 9: Might be worth noting that there is also `start_mongo.bat` for windows users.

ankhers (commented):
Overall this is fine. But there are a couple changes that would probably be beneficial.

.github/workflows/ci.yml
joeapearson: Oh excellent find.  Agreed.

.github/workflows/ci.yml
joeapearson: Alright.  Is there a reason not to want the continuous validation for every push?

.github/workflows/ci.yml
joeapearson: Yep.  Any suggestions for that?  I was starting simple so at least there is something in place as we go forward.

.github/workflows/ci.yml
joeapearson: I think this is probably fine for code qual stuff.  If it ever becomes a problem can fix then.

ankhers:
It is mostly because every push may not really need it. If you are
experimenting with some new feature and wanted to share some code or a
POC, it is less useful to show that tests are (maybe) failing.

But I will let you and Scott decide what is best for this repository. I
just wanted to share how I generally set things up.

CONTRIBUTING.md
joeapearson: Hmmm.. I don't see one.  Maybe there was in the past perhaps?

The contributing guide I've provided is hardly exhaustive, just meant to be a quick and dirty "get you going" for now.  In reality I've found you need to at least scan over `start_mongo.bash` and the tests to figure out what's going on.

ankhers:
A simple first step could be to decide the upper and lower bounds of
both Elixir and OTP versions you want to test against and just have the
matrix be those two sets. If you want to really make sure everything
works, you can install as many versions between those bounds that you
want to test against.

.github/workflows/ci.yml
joeapearson: Yes, more important here.  I'll fix.

ankhers:
This is fair. I believe there was one in the past but it may be gone
now. I think the number of people developing on windows is fairly low as
is.

joeapearson:
> A simple first step could be to decide the upper and lower bounds of both Elixir and OTP versions you want to test against

Yes agreed.  Again my preference is to start as basic as possible and then gradually expand our "official support" as we get used to taking care of this repo.  I started a discussion about that in #372 .

Thanks very much for the unexpected code review by the way, it was my understanding that you'd preferred to put the mongo work down for now otherwise I would have asked you directly.

joeapearson:
> This is fair. I believe there was one in the past but it may be gone now. I think the number of people developing on windows is fairly low as is.

I'm on a mac but when I have had to use Windows it's maddening to find that it's not been properly taken care of.  I think this should be a separate issue though.

ankhers:
> Thanks very much for the unexpected code review by the way, it was my
> understanding that you'd preferred to put the mongo work down for now
> otherwise I would have asked you directly.

My pleasure. Part of it is I am in the process of modifying the GH
actions at $DAYJOB, so I figured I could pitch in a bit here since it is
all still fresh in my head.

It's not that I am totally putting this project down, it is just that I
don't personally use mongo in any of my own projects at the moment, so I
may not be the best person to decide direction of the packages. However,
I am still more than happy to pitch in where it makes sense. I think
that is mostly in code reviews or if you or anyone else has have
questions about the code, I am more than happy to try and help out.

YasirAsar:
I'm so happy to see this repo active after long time. Thanks guys!!!

.github/workflows/ci.yml
scottmessinger:
```suggestion
          key: ${{ runner.os }}-{{runner.otp-version}}-{{runner.elixir-version}}-mix-${{ hashFiles('**/mix.lock') }}
```

I think this is what @ankhers  is suggesting!

scottmessinger:
@joeapearson I just cut a release! https://hex.pm/packages/mongodb/1.0.0-beta.1

The next step is to get this in before we hit 1.0.0.

etroynov:
@[joeapearson](https://github.com/joeapearson)

any news?

scottmessinger:
@etroynov It's shipped!

